PR #233, opened 2022-06-12
Allow custom & union productions

mccolljr:
This PR will eventually provide an implementation of "custom" productions (a custom parsing function that can parse any arbitrary go value), and "union" productions, which parse a single value from a set of possible values into a container of a single interface type.

The details regarding API and ergonomics will need to be fleshed out further before this can be merged.


context.go
```diff
@@ -94,13 +94,15 @@ func (p *parseContext) Stop(err error, branch *parseContext) bool {
 		p.deepestError = err
 		p.deepestErrorDepth = maxInt(branch.PeekingLexer.Cursor(), branch.deepestErrorDepth)
 	}
-	if branch.PeekingLexer.Cursor() > p.PeekingLexer.Cursor()+p.lookahead {
+	if !p.hasInfiniteLookahead() && branch.PeekingLexer.Cursor() > p.PeekingLexer.Cursor()+p.lookahead {
```
mccolljr, line 97: The impetus here is that sometimes, a grammar needs infinite lookahead. This becomes a bit more clear with the introduction of unions. See comments in `parser_advanced_test.go` for a more specific discussion of this

grammar.go
```diff
@@ -22,6 +22,38 @@ func newGeneratorContext(lex lexer.Definition) *generatorContext {
 	}
 }
 
+func (g *generatorContext) addUnionDefs(defs []unionDef) error {
+	unionNodes := make([]*union, len(defs))
+	for i, def := range defs {
```
mccolljr, line 27: Unions may rely on one another, so we have to add them all to the `typeNodes` map before we start processing their members, and then update the nodes in-place.

parser_advanced_test.go
```diff
@@ -0,0 +1,386 @@
...
+
+		withExprPrec1 = participle.UseUnion[ExprPrec1](
+			MemberBinAddSub{}, MemberBinMulDiv{}, MemberBinRem{}, MemberUnary{}, MemberIdent{}, MemberNumber{}, MemberString{}, MemberParens{})
+
+		withExprPrecAll = participle.UseUnion[ExprPrecAll](
+			MemberBinAddSub{}, MemberBinMulDiv{}, MemberBinRem{}, MemberUnary{}, MemberIdent{}, MemberNumber{}, MemberString{}, MemberParens{})
+	)
+
+	exprParser := mustTestParser(t, &Wrapper{},
+		participle.UseLookahead(-1), // this grammar requires infinite lookahead to parse correctly
```
mccolljr, line 303: This is the more detailed discussion I referred to earlier:

I wanted to see how it might work to use union productions for encoding a precedence hierarchy in a similar way to the custom parser from the other test in this file.

This particular grammar is allowed based on the implementation of `union` productions, but it will always fail to parse expressions that are longer than the lookahead length, because disambiguation can only happen when we determine the lowest-precedence operator in the chain, which necessarily means parsing a the expression before disambiguation can occur. (Additionally, this means that we will re-parse the expression in its entirety at each increasing precedence level until we find the one that is appropriate, which is a Performance Setback :tm:)

Obviously, this grammar can be amended to avoid this - this is 100% true. But nothing stops someone from writing this grammar, and nothing stops someone from trying to parse expressions with it. A grammar like this necessarily requires an infinite lookahead to be unambiguous.

Obviously, defaulting to infinite lookahead for unions could be very surprising, performance-affecting behavior. With this in mind, I chose, for this proof-of-concept implementation, to allow users to opt-in to infinite lookahead by passing a negative value to `UseLookahead`.

This is the "simplest" implementation of the `union` production type - it literally constructs an ad-hoc `disjunction` node to handle the parsing and then converts the results to the union's top-level type before return.

An alternative implementation, which is more work but could produce some serious performance gains, would be to borrow the "threaded" parsing implementation from https://swtch.com/~rsc/regexp/regexp3.html. I think, though, that doing this would require refactoring quite a bit of the parsing code, and I'm not sure if that's worthwhile. Perhaps this could be a future improvement.

I am open to suggestions around this.

options.go
```diff
@@ -41,6 +48,48 @@ func CaseInsensitive(tokens ...string) Option {
...
+			return fmt.Errorf("production parser must be a function (got %s)", parseFnType)
+		}
+		if parseFnType.NumIn() != 1 || parseFnType.In(0) != reflect.TypeOf((*lexer.PeekingLexer)(nil)) {
+			return fmt.Errorf("production parser must take a single parameter of type %s", peekingLexerType)
+		}
+		if parseFnType.NumOut() != 2 {
+			return fmt.Errorf("production parser must return exactly two values: the parsed production, and an error")
+		}
+		if parseFnType.Out(0).Kind() != reflect.Interface {
+			return fmt.Errorf("production parser's first return must be an interface type")
```
mccolljr, line 67: @alecthomas In the issue thread for #229 we discussed that this does not necessarily need to return an interface type. I'm flip-flopping back and forth on agreement with this.

On the one hand, in support of that conclusion: absolutely nothing about this code _requires_ that this type is an interface. The parse function could return anything at all.

On the other hand, against this conclusion:
What happens if someone provides a parsing function for `bool`, `Position`, `string`, `int64`, or any other type that the parser expects to have a specific meaning? I haven't tested it, but I suspect that this would lead to surprising behavior that could potentially be hard to track down.

It seems like there has to be some kind of limitation on the particular type(s) that can be parsed using a custom parsing function. `interface` types seem like a logical constraint, and I think it could even see that being extended to other complex types like structs, pointers-to-structs, slices-of-structs, and arrays-of-structs... but I'm not sure how to decide on a proper sub-set of types that are allowed.

It is _not_ a breaking change to accept more types, but if we have to remove types that are allowed, that _is_ a breaking change. That might be totally fine, as well - I'm of course open to whatever you think is the right initial approach - its just that my instinct is to be cautious: we could release the first version of this feature supporting only `interface` types, and then expand the allowed types in the future based on feedback from users who need other types here.

parser_advanced_test.go
```diff
@@ -0,0 +1,386 @@
...
+			MemberBinRem{}, MemberUnary{}, MemberIdent{}, MemberNumber{}, MemberString{}, MemberParens{})
+
+		withExprPrec2 = participle.UseUnion[ExprPrec2](
+			MemberBinMulDiv{}, MemberBinRem{}, MemberUnary{}, MemberIdent{}, MemberNumber{}, MemberString{}, MemberParens{})
+
+		withExprPrec1 = participle.UseUnion[ExprPrec1](
+			MemberBinAddSub{}, MemberBinMulDiv{}, MemberBinRem{}, MemberUnary{}, MemberIdent{}, MemberNumber{}, MemberString{}, MemberParens{})
+
+		withExprPrecAll = participle.UseUnion[ExprPrecAll](
+			MemberBinAddSub{}, MemberBinMulDiv{}, MemberBinRem{}, MemberUnary{}, MemberIdent{}, MemberNumber{}, MemberString{}, MemberParens{})
```
mccolljr, line 299: Perhaps this goes without saying, but the order of the arguments to the `UseUnion` function is important, since the parser will attempt to match them in the order given, and will return the first match. 

For example - If the arguments are provided in an order that will attempt to match `a` before attempting to match `a+b`, then it will always match the `a` in `a+b` and will never match `a+b` itself. In my opinion this behavior is ok as long as its documented, but wanted to call it out here in case there's a problem with this that I haven't thought about yet

options.go
alecthomas: Good thinking, I agree it's a reasonable place to start.

parser_advanced_test.go
```diff
@@ -0,0 +1,386 @@
...
+	exprParser := mustTestParser(t, &Wrapper{}, participle.UseCustom(parseExprAny))
+
+	requireParseExpr := func(src string, expected Wrapper) {
+		t.Helper()
+		var actual Wrapper
+		require.NoError(t, exprParser.ParseString("", src, &actual))
+		require.Equal(t, expected, actual)
+	}
+
+	requireParseExpr(`1`, Wrapper{ExprNumber{1}})
```
alecthomas, line 133: NBD but typically I would make this a table-driven test.

parser_advanced_test.go
```diff
@@ -0,0 +1,386 @@
...
+	"*": {3, 2},
+	"/": {5, 4},
+	"%": {7, 6},
+}
+
+type (
+	Expr interface{ expr() }
+
+	ExprIdent  struct{ Name string }
+	ExprString struct{ Value string }
```
alecthomas, line 29: Not `ExprString string`?

options.go
```diff
@@ -41,6 +48,48 @@ func CaseInsensitive(tokens ...string) Option {
...
+		if parseFnType.Out(1) != errorType {
+			return fmt.Errorf("production parser's second return must be %s", errorType)
+		}
+		prodType := parseFnType.Out(0)
+		p.customDefs = append(p.customDefs, customDef{prodType, parseFnVal})
+		return nil
+	}
+}
+
+func UseUnion[T any](members ...T) Option {
```
alecthomas, line 78: So beautiful, brings a tear to my eye 🥲

ebnf.go
```diff
@@ -51,6 +51,28 @@ func buildEBNF(root bool, n node, seen map[node]bool, p *ebnfp, outp *[]*ebnfp)
 			p.out += ")"
 		}
 
+	case *union:
```
alecthomas, line 54: 🙏

options.go
```diff
@@ -41,6 +48,48 @@ func CaseInsensitive(tokens ...string) Option {
 	}
 }
 
+func UseCustom(parseFn interface{}) Option {
```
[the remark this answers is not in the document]
alecthomas: Also this will need documentation.

options.go
alecthomas: Also this will need documentation.

parser_advanced_test.go
alecthomas: As I mentioned above, even a large one, having a finite lookahead is more about avoiding pathological cases. But I think it's fine as long as it's explicit opt-in.

re. optimisation, I've deliberately avoided doing anything too much in this area so as not to complicate maintenance of Participle. I was leaning towards putting all optimisations into code generation, though in practice the parser has continued to evolve (particularly with v2), and maintaining two parallel implementations with identical semantics would be non-trivial, so this hasn't happened in practice.

Also any optimisation in Participle's reflection-based parser will be minimal compared to code generation, so in addition to the extra maintenance complexity, it's probably time better spent there.

parser_advanced_test.go
```diff
@@ -0,0 +1,386 @@
...
+		Head ExprOperand        `@@`
+		Tail []MemberBinRemCont `@@+`
+	}
+
+	MemberBinRemCont struct {
+		Op   string      `@"%"`
+		Expr ExprOperand `@@`
+	}
+
+	ExprPrecAll interface{ exprPrecAll() }
```
alecthomas, line 229: This is super cool, love it!

It would be good to add (or replace the existing) this to the examples folder.

alecthomas (approved):
Absolutely love it, thanks for doing this!

I have no major objections, it all looks quite reasonable to me. I think the only two issues are lack of documents on the new functions (not unexpected for a draft!), and the name "custom" is not specific enough.

options.go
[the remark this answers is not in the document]
mccolljr: I wasn't sure if you wanted v2 to be backwards-compatible with pre-1.18 go versions but I guess the use of generics for the union precludes that a bit - I do like making this generic though, as it enforces+documents the signature on the function

parser_advanced_test.go
mccolljr: Ah tbh I wasn't really thinking about it, I was just declaring structs. I can make it a string!

options.go
```diff
@@ -21,6 +24,10 @@ func Lexer(def lexer.Definition) Option {
 //
 // Note that increasing lookahead has a minor performance impact, but also
 // reduces the accuracy of error reporting.
+//
+// If "n" is negative, this will be treated as "infinite lookahead".
```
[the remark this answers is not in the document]
mccolljr: Yeah this is reasonable - tbh I considered actually intercepting negative values inside the `UseLookahead` function and replacing them with `MaxIterations` for that reason - would you prefer that approach?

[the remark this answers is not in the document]
alecthomas: Not a bad idea, though I think `MaxIterations` is 1000000 which might exceed Go's stack depth. Perhaps another constant.

parser_advanced_test.go
mccolljr: Re: infinite lookahead - Agreed that arbitrarily large is better than truly infinite for real use cases, but for testing / troubleshooting I can see utility in being able to say "no, really, do this forever". If you're not opposed, I think I will document the risks of infinite lookahead and encourage using `99999` instead, but will leave the escape hatch there in case of "emergency", since it doesn't really cost anything to have it opt-in

Re: optimization - Heard. That makes sense, I'll leave this alone and just maybe document the performance risks in the example code

options.go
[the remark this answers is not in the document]
mccolljr: I think I'll leave the escape hatch there (could be useful in testing code, maybe - or to verify that a problem is indeed related to the lookahead value chosen), but will provide a constant that can be used for "really big but not actually infinite"

options.go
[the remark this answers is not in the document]
mccolljr: I'll go with `ParseTypeWith[T any]` and just keep the inner code the same.

options.go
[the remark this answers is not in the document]
mccolljr: I added some scare documentation to the infinite lookahead case, and defined a `MaxLookahead` constant that can be used for pseudo-infinite lookahead.

options.go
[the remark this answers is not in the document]
mccolljr: I changed this to `ParseTypeWith`, and I also changed the one for unions to `ParseUnion` to keep the naming consistent between the two.

parser_advanced_test.go
mccolljr: I changed my tests to be table-driven to be consistent with the rest of the codebase

parser_advanced_test.go
mccolljr: I was lazy and didn't change these, but I did move it to the `_examples/expr4` directory as an example of custom expression parsing.

options.go
mccolljr: Documented, also renamed to `ParseUnion` to be consistent with the `ParseTypeWith` option.

parser_advanced_test.go
mccolljr: I added this as `_examples/expr3` (and the custom parsing as `_examples/expr4`

mccolljr:
Updated naming and documentation, and converted the fat tests to examples. Also added some smaller tests to the main tests suite just to make sure we're covering the new functionality.

options.go
```diff
@@ -21,6 +29,14 @@ func Lexer(def lexer.Definition) Option {
...
 // Note that increasing lookahead has a minor performance impact, but also
 // reduces the accuracy of error reporting.
+//
+// If "n" is negative, it will be treated as "infinite" lookahead.
+// This can have a large impact on performance, and does not provide any
+// protection against stack overflow during parsing.
+// In most cases, using MaxLookahead will achieve the same results in practice,
+// but with a concrete upper bound to prevent pathological behavior in the parser.
+// Using infinite lookahead can be useful for testing, or for parsing especially
+// ambiguous grammars. Use at your own risk!
```
alecthomas, line 39: 👍

options.go
```diff
@@ -41,6 +57,57 @@ func CaseInsensitive(tokens ...string) Option {
...
+//    U = X | Y | Z | W .
+// When the parser encounters a field of type T, it will attempt to parse each member
+// in sequence and take the first matche. Because of this, the order in which the
+// members are defined is important. You must be careful to order your members appropriately.
+//
+// An example of a bad parse that can happen if members are out of order:
+//
+// If the first member matches A, and the second member matches A B,
+// and he source string is "AB", then the parser will only match A, and will not
+// try to parse the second member at all.
```
alecthomas, line 95: 🙏

alecthomas (approved):
Again, awesome, thanks for implementing this :)

mccolljr:
This is awesome, thanks for coordinating with me to get this merged so quickly!

mccolljr:
> Thanks for doing the work :)
> 
> PS. I updated the README, and made one minor change: renamed the `ParseUnion` option to `Union`, and tagged a new alpha release.

Saw that - that's a fine change, naming is hard 😅

PR #458, opened 2026-05-24
fix: avoid panic in Unquote on tokens shorter than two characters

SAY-5:
`unquote` indexes `s[0]` and slices `s[1:len(s)-1]` without checking the length first, so a token whose value is empty or a single character (which can reach the `Unquote` mapper depending on the lexer rules) panics with an index/slice out-of-range instead of returning the usual parse error.

This adds a length guard that returns an error for tokens too short to be a quoted literal, matching the existing error path that wraps into a parse error. Added a regression test that feeds a single-quote token through `Unquote` and asserts an error is returned rather than a panic.

### DIFF
--- a/_examples/go.mod
+++ b/_examples/go.mod
@@ -1,6 +1,6 @@
 module github.com/alecthomas/participle/v2/_examples
 
-go 1.18
+go 1.24
 
 require (
 	github.com/alecthomas/assert/v2 v2.11.0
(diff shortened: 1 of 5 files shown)


alecthomas:
I think it'd be good to have a more useful error, ideally with positional information, but this is not a parse error - this is a programmer error and should be a panic.

SAY-5:
Good point, this is a programmer error rather than a parse error. Updated to panic instead.

alecthomas (approved):

alecthomas:
Thanks!

alecthomas:
Thanks!

PR #469, opened 2026-08-09
feat: add fuzzing example and documentation

Pastalikek65:
Partially addresses #204

Grammar-driven generation (`parser.Fuzz(&ast)` producing random *valid* strings) would be a larger project — it needs a regex-to-string generator in the lexer to invert captures. This PR instead delivers the immediately useful half of the request: an official, ready-to-run native Go fuzz target for participle parsers.

Changes:
- `fuzz_test.go`: `FuzzParse` — a concrete example target that exercises a grammar with arbitrary input (panic/hang detection), with a seed corpus that also runs in CI as a regular test
- `README.md`: new "Fuzzing" section explaining the pattern and usage (`go test -fuzz=FuzzParse -fuzztime 30s ./...`)

If grammar-driven generation is still desired, it would fit well as a separate issue with a design for a `lexer.Pattern` → string generator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,6 +584,27 @@ thrift takes 630ms, which aligns quite closely with the benchmarks.
 
 A compiled `Parser` instance can be used concurrently. A `LexerDefinition` can be used concurrently. A `Lexer` instance cannot be used concurrently.
 
+## Fuzzing
+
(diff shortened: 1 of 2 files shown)


alecthomas:
@codex review

PR #470, opened 2026-08-09
fix: inject Pos/EndPos/Tokens into Capture and TextUnmarshaler values

### LINKED ISSUE #132: Regression on #35
It seems like, with the `v2.0.0-alpha3` version of participle, positions are, again, no longer added to structs that implement the `Capture` interface. This apparently was first fixed in f14a17d in response to #35, but the associated test has since been removed and the issue has reappeared.

I have also tested that this issue is present on `master` by adapting and adding the old `TestIssue35` into `parser_test.go` and observing a failed test run.

Pastalikek65:
Closes #132

`Pos`/`EndPos`/`Tokens` are injected automatically for struct nodes parsed via `@@`, but user-defined capture targets implementing `Capture` or `encoding.TextUnmarshaler` never had them populated — the `Pos lexer.Position` field stayed zero (verified on master with `@Ident @Ident`).

Changes:
- `nodes.go`: new `applyCapturePosition` helper populates `Pos`, `EndPos` and `Tokens` on capture targets (guarding non-struct types such as `type count int`), called from the `Capture` and `TextUnmarshaler` branches of `setField`, and per-element for `[]Capture` slices.
- `parser_test.go`: `TestCaptureStructPosition` covering `Pos`, `EndPos` and `Tokens` on a `Capture` struct.

`EndPos` is computed as the last captured token's position advanced by its length (byte offset, rune column) — an approximation of the next-token position the struct path uses; it doesn't account for embedded newlines in captured values.

### DIFF
--- a/nodes.go
+++ b/nodes.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/alecthomas/participle/v2/lexer"
 )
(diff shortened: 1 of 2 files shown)


alecthomas:
@codex review

PR #477, opened 2026-08-10
perf(parser): reduce allocations in the parse hot path

Pastalikek65:
Reduces allocations in the parser hot path.

## Results (microc benchmark)

| metric | before | after | delta |
|---|---|---|---|
| allocs/op | 59694 | 55054 | **-7.8%** |
| B/op | 4572762 | 4161671 | **-9.0%** |

ns/op was noisy on the test machine; the allocation/byte reductions are stable and directly reduce GC pressure, which is the dominant cost on this benchmark's CPU profile (mallocgc ≈ 27% of samples).

## Changes

1. **`apply []contextFieldSet` by value** — previously `[]*contextFieldSet`, which heap-allocated a `contextFieldSet` per capture. Now captures are stored inline in the slice; the only allocation is the amortised slice growth.

2. **`contextFieldSet.field` is now `*structLexerField`** — the field comes from the immutable grammar node owned by the parser for its whole lifetime, so a pointer is safe. Shrinks `contextFieldSet` from ~200B to ~88B, halving the cost of slice-growth copies (`Defer` was 5.5k allocs/op).

3. **Decouple the optional trace from `parseContext`** — new `traceState` holds the writer and depth, allocated only when `Trace` is enabled. The deferred exit closure now captures only `traceState`, never the parse context, and the now-unused `depth` field is removed.

## Notes on what I looked at and didn't do

The largest remaining allocation is the ~10.8k/op of `parseContext` branch copies in `disjunction`/`group`. These escape to the heap because `parseable` and `custom` nodes hand `&ctx.PeekingLexer` to user code / `reflect.Call`, which forces Go's escape analysis to treat every `node.Parse(ctx, ...)` interface call as potentially retaining `ctx`. Eliminating those allocations would require pooling or API changes that risk incorrect behaviour if user code retains the lexer, so they're out of scope here.

## Validation

- `go test ./...` passes (full suite, including the trace tests via `participle.Trace`).
- `go vet` clean for the changed package (pre-existing struct-tag warnings in tests remain).
- `gofmt` clean.

### DIFF
--- a/context.go
+++ b/context.go
@@ -12,20 +12,20 @@ import (
 type contextFieldSet struct {
 	tokens     []lexer.Token
 	strct      reflect.Value
-	field      structLexerField
+	field      *structLexerField
 	fieldValue []reflect.Value
 }
 
 // Context for a single parse.
 type parseContext struct {
 	lexer.PeekingLexer
-	depth             int
 	trace             io.Writer
+	traceState        *traceState
 	deepestError      error
 	deepestErrorDepth int
 	lookahead         int
 	caseInsensitive   map[lexer.TokenType]bool
-	apply             []*contextFieldSet
+	apply             []contextFieldSet
 	allowTrailing     bool
 }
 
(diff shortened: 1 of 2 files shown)


alecthomas:
@codex review

PR #483, opened 2026-09-16
perf: avoid per-token mapper slice allocation

srinathshrestha:
Follow-up to [https://github.com/alecthomas/participle/discussions/482](https://github.com/alecthomas/participle/discussions/482).

The mapping lexer currently creates a temporary `combined []Mapper` for each token, appending the global & token-specific mapper slices. This PR changes it to iterate over those existing slices directly. The mapper order, error short-circuiting, original-token-type selection, and existing EOF behavior are covered by tests.

Isolated mapping-wrapper benchmark on my machine, using an allocation-free lexer with one global and one token-specific identity mapper:

```yaml
before: ~86.4 ns/op    16 B/op    1 alloc/op
after:  ~64.6 ns/op     0 B/op    0 allocs/op
```

These are median results from 10 runs. This benchmark isolates mapper dispatch and is not intended to represent whole-parser performance.

All existing tests, the new focused tests, race tests, example tests, and the configured linter pass.


alecthomas (approved):

alecthomas:
Thanks!

